PR #31, opened 2019-05-28
Move pressure to Nek node. Refs #15

aprilnovak:
This MR moves `pressure` to the Nek node since it is the only tool that requires pressure. If we add subchannel capabilities to `SurrogateHeatDriver`, we can move pressure back up.


RonRahaman:
I'll implement this in #26 .  Thanks!

PR #60, opened 2019-08-06
Integer types in Nek interface, drivers

RonRahaman:
This is a second pass at #54.  The changes can be summarized as follows:
* The Nek interface (C and Fortran) has been updated to use `int32_t` for indices.  The submodule has been updated accordingly.  
* In ENRICO, loop indices that are used to index Nek arrays are declared as `int32_t`.  `gsl::index` isn't really appropriate here, since the Nek arrays can only be as long as an int32.  
* Loop indices that are used to index `std::vector` and xtensors are declared as `gsl::index`

### DIFF
--- a/include/enrico/nek_driver.h
+++ b/include/enrico/nek_driver.h
@@ -61,7 +61,7 @@ public:
   //!
   //! \param local_elem The local index of the desired element
   //! \return The dimensionless coordinate of the element's centroid
-  Position centroid_at(int local_elem) const;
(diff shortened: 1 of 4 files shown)


RonRahaman:
@paulromano I believe the accepted practice in Nek is to compile with `-fdefault-integer-8` .  Fun!  It does raise questions about what to do in ENRICO when we start running large problems.  We might have to make use of some preprocessor macros.  I'd like to tackle that later.

PR #98, opened 2020-06-01
Updates to ENRICO build process to allow for optional nek build

lebuller:
This PR addresses #92 by making adjustments to the CMake build files, and adding a check that ENRICO is built with support for Nek5000 at runtime.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,12 @@
 cmake_minimum_required(VERSION 3.3)
 project(enrico Fortran C CXX)
 
+option(NEK_BUILD  "Determines if Nek5000 will be included" ON)
+
+if (NEK_BUILD)
+    add_definitions(-DNEK_BUILD)
+endif()
+
 # On BG/Q, linking requires -dynamic, not -rdynamic
 if (CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "ppc64")
   set(CMAKE_SHARED_LIBRARY_LINK_CXX_FLAGS "-dynamic")
@@ -20,15 +26,17 @@ endif ()
 # =============================================================================
 # Discover name-mangling for routines in libnek5000
 # =============================================================================
-include(FortranCInterface)
-FortranCInterface_VERIFY()
-FortranCInterface_HEADER(${CMAKE_BINARY_DIR}/nek_mangling.h
-    MACRO_NAMESPACE C2F_
-    SYMBOL_NAMESPACE C2F_
-    SYMBOLS
-    nek_init
-    nek_end
-    nek_solve)
+if (NEK_BUILD)
+    include(FortranCInterface)
+    FortranCInterface_VERIFY()
+    FortranCInterface_HEADER(${CMAKE_BINARY_DIR}/nek_mangling.h
+        MACRO_NAMESPACE C2F_
+        SYMBOL_NAMESPACE C2F_
+        SYMBOLS
+        nek_init
+        nek_end
+        nek_solve)
+endif()
 
 # =============================================================================
 # Headers for all targets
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -101,9 +115,13 @@ add_library(libenrico
             ${SOURCES})
 
 set_target_properties(libenrico PROPERTIES OUTPUT_NAME enrico)
-target_link_libraries(libenrico PUBLIC iapws libnek5000 libopenmc xtensor heat_xfer pugixml gsl-lite)
+target_link_libraries(libenrico PUBLIC iapws libopenmc xtensor heat_xfer pugixml gsl-lite)
```
pshriwise, line 118: I'm thinking it would be good to set a variable for the list of libraries found here (following with the `SOURCES` variable, maybe `LIBRARIES` could be used) and add `libnek5000` to that list conditionally depending on the state of `NEK_BUILD`. This would allow us to have only one call to `target_link_libraries` and makes the lines a little more readable IMO.

`target_link_libraries(libenrico PUBLIC ${ENRICO_LIBS})`

src/coupled_driver.cpp
```diff
@@ -121,7 +125,11 @@ CoupledDriver::CoupledDriver(MPI_Comm comm, pugi::xml_node node)
   // Instantiate heat-fluids driver
   std::string s = heat_node.child_value("driver");
   if (s == "nek5000") {
+#ifdef NEK_BUILD
     heat_fluids_driver_ = std::make_unique<NekDriver>(heat_comm.comm, heat_node);
+#else
+    throw std::runtime_error{"nek5000 driver selected but not ON in build options"};
```
pshriwise, line 131:
```suggestion
    throw std::runtime_error{"Nek5000 was specified as a solver, but is not enabled in this build of ENRICO."};
```

CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 cmake_minimum_required(VERSION 3.3)
 project(enrico Fortran C CXX)
 
+option(NEK_BUILD  "Determines if Nek5000 will be included" ON)
```
RonRahaman, line 4: Could you change this to `USE_NEK`?  That way, it will match the the `USE_SHIFT` variable from #97

CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
...
+
+if (NEK_BUILD)
+    add_definitions(-DNEK_BUILD)
```
RonRahaman, line 7: I would prefer `target_compile_definitions`

CMakeLists.txt
lebuller: So replace all current `target_link_libraries` calls with with setting or appending to a `LIBRARIES` variable and then link all libraries with one call? And will the location of `target_compile_definitions` for `libenrico` need to be changed to after the single `target_link_libraries` call?

CMakeLists.txt
```diff
@@ -132,10 +147,19 @@ set_target_properties(
...
         iapws
-        test_nek5000_singlerod
         test_openmc_singlerod
         PROPERTIES CXX_STANDARD 14 CXX_EXTENSIONS OFF)
 
+
+if (NEK_BUILD)
+    add_executable(test_nek5000_singlerod tests/singlerod/short/test_nek5000.cpp)
+    target_link_libraries(test_nek5000_singlerod PUBLIC libenrico)
+
```
pshriwise, line 168:
```suggestion
```

CMakeLists.txt
```diff
@@ -101,14 +111,33 @@ add_library(libenrico
...
 target_compile_definitions(libenrico PRIVATE GSL_THROW_ON_CONTRACT_VIOLATION)
 
+if (USE_NEK)
+    target_compile_definitions(libenrico PUBLIC USE_NEK)
+endif()
+
+if (USE_NEK)
+    list( APPEND LIBRARIES
+        libnek5000)
+endif()
```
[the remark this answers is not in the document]
pshriwise: Do you have any suggestions for the naming the compiler definition @RonRahaman? I was thinking maybe `NEK5K` or simply `NEK`?

pshriwise:
Thanks @lebuller!

PR #112, opened 2020-08-28
Temperature_ic, density_ic vals from XML

RonRahaman:
In the docs for `enrico.xml`, the allowed values for `<temperature_ic>` and `<density_ic>` were listed incorrectly.  The allowed value for the heat/fluid driver should be "heat_fluids" instead of "heat".

### DIFF
--- a/doc/source/input.rst
+++ b/doc/source/input.rst
@@ -191,7 +191,7 @@ value of "robbins-monro" indicates that Robbins-Monro relaxation is to be used:
 The initial temperature distribution can be determined either from the
 neutronics solver or the heat-fluids solver. A value of "neutronics" will use
 the temperatures specified in the model for the neutronics solver whereas a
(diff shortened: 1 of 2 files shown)


RonRahaman:
There was a second issue in the code itself.  `xml::node::child_value` returns a `const char_t*`, and the subsequent comparisons failed.  In this PR, I now declare `s` as a `std::string`, which performs the conversion and allows the comparisons to succeed.  Elsewhere, this conversion was done properly.  
```
  if (coup_node.child("temperature_ic")) {
    auto s = coup_node.child_value("temperature_ic");

    if (s == "neutronics") {
      temperature_ic_ = Initial::neutronics;
    } else if (s == "heat_fluids") {
      temperature_ic_ = Initial::heat;
    } else {
      throw std::runtime_error{"Invalid value for <temperature_ic>"};
    }
  }
```

PR #114, opened 2020-10-29
Add more detailed instructions for building.

aprilnovak:
I added what I believe are fully detailed instructions for building Enrico with OpenMC and the various T/H solvers (some minor things were missing, like `-DNEK_DIST=none`, and unzipping the restart file for the Nek5000 example.

Just like to be as explicit as possible!


RonRahaman (approved):
Looks great, thanks!

PR #115, opened 2020-12-10
Update NekRS submodule; run nrspre in CI

RonRahaman:
This updates the NekRS submodule to pull specific commit hashes of OCCA and Nek5000.  Upstream NekRS pulls OCCA and Nek5000 from their respective master branches at configure time; however, this introduced some version skew that broke the builds.  By pulling from specific commit hashes (rather than branches) we should be able to avoid this version skew.

### DIFF
--- a/ci/test_singlerod_openmc_nekrs.sh
+++ b/ci/test_singlerod_openmc_nekrs.sh
@@ -3,4 +3,5 @@ set -ex
 
 cd tests/singlerod/short/openmc_nekrs
 export NEKRS_HOME=$(realpath ../build/install)
+../build/install/bin/nrspre rod_short 2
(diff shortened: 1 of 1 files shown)


aprilnovak:
Looks good! Just curious about the `nrspre` change though - why was this added?

aprilnovak:
My question was just a curiosity thing, so I'll go ahead and merge this so that you can move forward!

RonRahaman:
Thanks, April!  I found that the Travis builds failed if I didn’t pre-compile the NekRS kernels, hence the nrspre.  NekRS needs the pre-compile on a lot of different machines, so I’m not concerned that we need it on the Travis servers, too.

PR #122, opened 2021-01-26
Use cells rather than cell handles in Shift calls.

sphamil:
Corrects use of cell index in calls to Shift functions. The ENRICO cell handle was being passed instead of the Shift cell index. Also changes the variable names to reflect the distinction between handles and cell indices.


paulromano (approved):
Thanks for the fix!

PR #126, opened 2021-02-18
WIP: nekRS update 2021-02-15

RonRahaman:
Diagnostic PR for upstream nekRS updates.  Currently, the coupled OpenMC/nekRS shortrod cases fail during nekRS solve:
```
step= 1  t= 1.00000000e-04  dt=1.0e-04  C= 4.84  UVW: 8  P: 18  S: 94  eTime= 4.31e-01, 4.31334e-01 s
step= 2  t= 2.00000000e-04  dt=1.0e-04  C= 31.02  UVW: 8  P: 24  S: 57  eTime= 4.18e-01, 8.49406e-01 s
Unreasonable CFL! Dying ...
```

In a standalone nekRS, the shortrod problem has similar results:
```
step= 1  t= 1.00000000e-04  dt=1.0e-04  C= 4.84  UVW: 8  P: 18  S: 1  eTime= 2.33e-01, 2.33210e-01 s
step= 2  t= 2.00000000e-04  dt=1.0e-04  C= 31.02  UVW: 8  P: 24  S: 1  eTime= 2.92e-01, 5.24758e-01 s
Unreasonable CFL! Dying ...
```

PR #127, opened 2021-02-22
Fix a bug in recognizing boundary condition

yslan:


yslan:
This might help the PR  #126

RonRahaman:
Thanks, @yslan !  That did the trick.  
Closes #126

PR #130, opened 2021-03-16
MPICH_FC in GitLab CI

RonRahaman:
The `ubuntu-latest` in the GitLab CI runners have swtiched from Ubuntu 18 to 20.  The MPICH distribution in Ubuntu 20 requires the evn variable `MPICH_FC=gfortran` to configure Nek5000 properly.  This PR adds that env variable.


paulromano (approved):
Thanks for figuring this out!

PR #140, opened 2021-06-04
Remove wrong barrier in NekRSDriver

RonRahaman:
This removes an incorrect barrier in `NekRSDriver`.   It fixes coupled neutronics/nekRS runs with disjoint communicators.  

The nekRS driver's comm was passed to the barrier function.  However, the barrier function is called by all the ranks in the coupled driver's comm, including the ranks that are not active in the nekRS driver's comm.  In runs with disjoint comms, this causes MPI errors due to an invalid comm in the `MPI_Barrier` call.


paulromano (approved):
Love it when the fix is this simple. Thanks @RonRahaman!

PR #142, opened 2021-06-11
localq as passive scalar in Nek5000

RonRahaman:
In this PR, the local heat source (previously `localq`) is stored as an (unsolved) passive scalar in Nek5000.  This allows localq to be easily output in the Nek5000 fld file.

This requires a few changes in the way we setup the Nek problems.  
* `ldimt` in `SIZE` (the amount of storage for passive scalars, including temperature) **must** be >= 2.  That is, there must be at least space for both temperature and localq.  This is verified in `Nek5000Driver` and will raise an error if not satisified.
* p023 in the parfile (the number of solved passive scalars, excluding temperature) **must** be < `ldimt - 1`.  That is, there must be an unsolved scalar.  This is also verified in `Nek5000Driver`

It is also recommended (but not required) that the .rea file specifies no I/O by setting p014 and p015 to a negative values.  This is because file I/O is controlled by Nek5000Driver::write_step, which writes I/O using `nek_write_step` from the ENRICO branch of Nek5000:
https://github.com/enrico-dev/Nek5000/blob/e7edb50a18b272015bec9ea8946267ce69e85b47/core/nek_interface.f#L268-L279

The resultant .fld files contain localq as the last passive scalar.  The units are W/cm^3.


src/nek5000_driver.cpp
```diff
@@ -143,6 +155,10 @@ int Nek5000Driver::set_heat_source_at(int32_t local_elem, double heat)
   return nek_set_heat_source(local_elem + 1, heat);
 }
 
+void Nek5000Driver::write_step(int timestep, int iteration) {
+  nek_write_step();
```
RonRahaman, line 159: This is implemented here:
https://github.com/enrico-dev/Nek5000/blob/e7edb50a18b272015bec9ea8946267ce69e85b47/core/nek_interface.f#L268-L279

hapfang (commented):
My understanding is that ENRICO now saves the heat source term by default, and it will report an error if `ldimt` < 2. Right? Can we still keep the option not to write out the heat source? After all, the heat source is helpful only for testing purposes. We probably do not need it for production runs. Also, if we are doing the ENRICO simulations for a large case like the full core, one extra field in the restart files will take up a lot of disk space.

RonRahaman:
Sure, that's a good idea.  I can an implement an `<output_heat_source>` option to the enrico.xml file, which will be "false" by default.

PR #143, opened 2021-06-17
Output heat source with NekRSDriver (revised)

RonRahaman:
This PR allows the `NekRSDriver` to output local heat source to a Nek field file.  

It works as follows:
* The separately-allocated `localq` vector is wrapped as an `occa::memory` object and passed to `writeFld`.  By using `occa::cpu::wrapMemorry`, we do not need to copy the contents of `localq`.  
* The heat source is output to a new field file prefixed with `qsc`, where the heat source is in the place of the temperature field.  
* The units are W/cm^3

This capability is disabled by default.  It can be enabled by setting`<output_heat_source>false</output_heat_source>` in the `<heat_fluids>` block of `enrico.xml`.


doc/source/conf.py
```diff
@@ -104,5 +104,5 @@ def build_doxygen(app):
 # -- Setup hooks -------------------------------------------------------------
 
 def setup(app):
-    app.add_stylesheet('theme_overrides.css')
+    app.add_css_file('theme_overrides.css')
```
RonRahaman, line 107: `add_stylesheet` was deprecated in favor of `add_css_file`.  See, for example: https://github.com/sphinx-doc/sphinx/issues/7747

hapfang (approved):

PR #158, opened 2021-11-03
Remove gather functions from HeatFluidsDriver

RonRahaman:
This PR removes the global gather functions from `HeatFluidsDrivers`.  These functions are not used in ENRICO's current coupling scheme.  

In ENRICO's earliest coupling scheme, the heat/fluids driver's local fields were gathered into global fields on one neutronics rank.  This was feasible for small models, but it's totally infeasible for running large-scale problems.  The current coupling scheme never gathers global fields from the heat/fluid driver, and I strongly believe that no future implementation should consider it a viable option.  

The changes to `HeatFluidsDriver` are:
* The functions that return the global fields (`centroids`, `volumes`, `temperature`, `density`, and `fluid_mask`) are removed.
* The functions that return the local fields (`temperature_local`, `fluid_local`, etc.) are renamed (to `temperature`, `fluid`, etc.), since it's no longer necessary to disambiguate them from the functions that return the global fields
* The definition of `init_displs` is removed.  That was used to determine displacements for the global gather.  After removing the global gathers, it is not needed.

The changes to `Nek5000Driver`, `NekRSDriver`, and `SurrogateHeatDriver` are:
* Renaming of functions that return local fields, as noted above.  

The changes to `Comm` are:
* The definition of `Gatherv` is removed.


paulromano (approved):
Always love to see PRs with net negative LOCs!

PR #159, opened 2021-11-03
Update docs:  Comm Pattern for solution transfer

RonRahaman:
This updates the docs with some diagrams that show the communcation pattern for the temperature/density updates and the heat source update.


paulromano (approved):
The new figures are great -- thanks once again @RonRahaman!

Issue #152, opened 2021-10-12
Doesn't use external libopenmc.so and instead installs conflicting libopenmc.so


yurivict:
This causes conflict with pre-installed OpenMC.

Please use external OpenMC when it is present.

yurivict:
This would still conflict with the standalone OpenMC package because ```libopenmc.so``` would be installed twice. Same library can't be installed twice because ldconfig resolved its name into path.

You can link static libraries into dynamic libraries fine as long as they are built with -fPIC.

RonRahaman:
Are you able to successfully install via something like this?  
* I'm inferring the MPI compiler paths from the CMake command you quoted above.  
* I excluded these all the explicit include paths, link paths, libs, and RPATH flags.  CMake's compiler verifications, `find_package(MPI)`, and the way we set the RPATH should be sufficient.  
* I've specified set `CMAKE_INSTALL_PATH` to `/usr/local/enrico`, so your systemwide openmc isn't clobbered
* I've retained your opt flags and other CMake options
```
cmake \
-DNEK_DIST=none \
-DCMAKE_C_COMPILER:STRING="/usr/local/mpi/openmpi/bin/mpicc"  \
-DCMAKE_CXX_COMPILER:STRING="/usr/local/mpi/openmpi/bin/mpicxx"  \
-DCMAKE_C_FLAGS:STRING="-O2 -pipe -fno-omit-frame-pointer  -fstack-protector-strong -fno-strict-aliasing " \
-DCMAKE_C_FLAGS_DEBUG:STRING="-O2 -pipe -fno-omit-frame-pointer  -fstack-protector-strong -fno-strict-aliasing "  \
-DCMAKE_C_FLAGS_RELEASE:STRING="-O2 -pipe -fno-omit-frame-pointer  -fstack-protector-strong -fno-strict-aliasing "  \
-DCMAKE_CXX_FLAGS:STRING="-O2 -pipe -fno-omit-frame-pointer -fstack-protector-strong -fno-strict-aliasing -fno-omit-frame-pointer "  \
-DCMAKE_CXX_FLAGS_DEBUG:STRING="-O2 -pipe -fno-omit-frame-pointer -fstack-protector-strong -fno-strict-aliasing -fno-omit-frame-pointer "  \
[...]

PR #163, opened 2022-03-30
Use installed OpenMC if available

kkiesling:
The goal of this change is to find OpenMC if it is already installed and use that instead of the submodule. Updates also include a change to the `gsl` headers that are included throughout the code. That is both a best practice now and also necessary for this change to work.

I believe this should address #152.


paulromano (approved):
Thanks @kkiesling!

PR #168, opened 2022-07-19
updated catch to latest stable v2.x version

### LINKED ISSUE #167: update vendor repo Catch2 to latest release
The vendored repo `catch` ([Catch2](https://github.com/catchorg/Catch2/tree/ab80277a861276ca0e5ad5062597c642da4a6933)) currently sits at a version of roughly 2.9.2, last updated in 2019, where as the latest release is 3.1.0. The current version being used here is not compatible with the M1 chip in new Macs, but that appears to be fixed in newer versions of Catch2. However, `catch.hpp` no longer exists with these newer versions, so headers in a few test files here will also need to be updated (see [release documentation](https://github.com/catchorg/Catch2/releases/tag/v3.0.1)).

kkiesling:
fixes #167 

I updated the catch2 repo to the latest stable version (v2.x branch). The latest stable version is compatible with M1 but does not require any updates on our end.


paulromano (approved):
Simple enough -- thanks @kkiesling!

PR #170, opened 2022-11-28
ENRICO nekRS_driver Update to Support nekRS v22

hapfang:
This MR contains changes to `nekrs_driver.cpp` to make ENRICO compatible with nekRS v22. The updated ENRICO has been tested to run successfully on both local CPU clusters and ORNL Summit. 

**Caveats**: 
 - The current nekRS driver also requires necessary bug fixes within the nekRS source, which has not been merged in the latest nekRS repo yet. A follow-up update will be committed once the [nekRS github repo](https://github.com/Nek5000/nekRS) is updated 
 - The updated ENRICO not necessarily shows a performance boost for small-scale simulations (e.g., a single rod test case). Larger scale simulations are required to see the benefits of performance enhancement from nekRS CFD solver.


hapfang:
This PR is no longer needed with the new PR that provides nekRS v23 support.

